PR #182, opened 2024-02-20
Feat/article-3-collections-and-news

JohanHjelsethStorstad:
This PR creates
- news(articles)
- article categories

changes:
- relations between cms things to make cascading work probably
- Some small bugs here and there
- Changes to follow conventions


Paulijuz (approved):

Paulijuz:
This looks like the interesting part.

```
#32 40.85 Invalid `prisma.articleCategory.findMany()` invocation:
#32 40.85 
#32 40.85 
#32 40.85 error: Environment variable not found: DB_URI.
#32 40.85   -->  schema.prisma:16
#32 40.85    | 
#32 40.85 15 |   provider = "postgresql"
#32 40.85 16 |   url      = env("DB_URI")
#32 40.85    | 
#32 40.85 
```

It appears that during the build prisma tries to connect to the db, but it fails since a DB_URI is not set during build...

JohanHjelsethStorstad:
I need someone else to look over it all @theodorklauritzen, @AndreasVJ.

theodorklauritzen:
Things look pretty nice, but I noticed a few weird things.
- Since names cannot be duplicated on news, they will probably be long. Therefore, the character limit of 20 is too small. Another option is to include the publishing year or date in the URL and then allow equal names on news articles.
- When editing news, the title and name fields should sync. Or the name field should not be there.

theodorklauritzen (approved):
LGTM

JohanHjelsethStorstad:
> Things look pretty nice, but I noticed a few weird things.
> 
> * Since names cannot be duplicated on news, they will probably be long. Therefore, the character limit of 20 is too small. Another option is to include the publishing year or date in the URL and then allow equal names on news articles.
> * When editing news, the title and name fields should sync. Or the name field should not be there.

I have fixed unique name: now: a news article has @@unique([articleName, orderPublished]) and an article in a article category must be the only article with that name in the category

Paulijuz (approved):
Looks good!

PR #183, opened 2024-02-21
fix/navigation when updating and creating an article was not correct

JohanHjelsethStorstad:
I forgot to change the navigation when updating and creating a news article in the last pr. The URL used to be news/[name], now it is news/[order]/[name] since name is no longer unique


Paulijuz (approved):
Looks good!

PR #193, opened 2024-03-05
Feat/ombul

JohanHjelsethStorstad:
This branch adds most of the ombul functionality needed
- Refactors how to upload things to the store in a standard way
- Changes how image collections works so It has special collections
- All ombul covers are saved to a special collection
- Special collections will be authed by permissions not the upcoming visibility system
- Will later implement that a cmsImage can only point to a Collection with certain permission (READ_OMBUL)
- Implements a PdfDocument component to view imbuls in a nice way
- Implements all CRUD for ombuls with corresponding permissions


src/app/ombul/[...yearAndName]/page.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const ombulRes = await readOmbul({
+        name,
+        year
+    })
+    if (!ombulRes.success) notFound()
+    const ombul = ombulRes.data
+
+    const path = `/store/ombul/${ombul.fsLocation}`
+
+    const user = await requireUser({
```
Comment (theodorklauritzen) [inline src/app/ombul/[...yearAndName]/page.tsx:32]: This should be called before readOmbul. ReadOmbul will throw an error and then notFound will be called, before we can check this.

src/actions/ombul/destroy.ts
```diff
@@ -0,0 +1,47 @@
+'use server'
+import prisma from '@/prisma'
+import errorHandler from '@/prisma/errorHandler'
+import deleteFile from '@/store/deleteFile'
+import type { ActionReturn } from '@/actions/Types'
+import type { ExpandedOmbul } from './Types'
+
+export async function destroyOmbul(id: number): Promise<ActionReturn<ExpandedOmbul>> {
```
theodorklauritzen, line 8: Permissions?

theodorklauritzen (approved):

PR #197, opened 2024-03-07
Feat/migrations-DobbelOmega

JohanHjelsethStorstad:
The basics for writing migrations and some migrations.
Infrastructure this PR includes
- The ability to have two schemas and clients
- Ability to generate veven client and pull schema from. veven
- Import aliases for prismaservice using tscpaths
- A standard way to set up a migration
- ENV vars to control behaviour of DobbelOmega
- A system for limiting migrations including "mapping" of ids (IdMapper)

Migrations:
- ImageCollections (ImageGroups on veven)
- Images - including fetching indifferent sizes and DB row
- Ombuls - including fetching of pdf and linking to correct coverimage
- Omegaquotes
- (news)articles - all Articles (displayed on /news on veven) migrated as a NewsArticle (with relation to a Article)

Still missing:
- Needless to say: many more migrations
- A logger so we have a log so we can actually remember the final time we DobbelOmega-ed

And this system is documented in /docs.


Paulijuz (approved):
Cannot be bothered to test as it will take a long time, but i have seen that it works so it's probably fine. Looks good!

PR #198, opened 2024-03-08
Chore/change order published to relation

JohanHjelsethStorstad:
I think it probably makes more sense to have things that use order (only newsArticles for now) have a relation to omegaOrder  that references order rather than copying. It is a very minor change, but I think it from a sql modeling perspective makes more sense.


Paulijuz (approved):
Looks good!

PR #200, opened 2024-03-11
Refactor/special images and cms

JohanHjelsethStorstad:
This PR is meant to implement the special pattern on cms and images in the same way as for imageCollections.
This PR removes:
- CmsImages will no longer be created if they don't exist on read
- CmsParagraphs will no longer be created if they don't exist on read

This PR changes:
- CmsImage now take in the CmsImage (so no dobbel fetching is needed in articles for example) (instead the new SpecialCmsImage behaves in the old way by calling readSpecialCmsImage).
- A new pattern to write long reusable include statements.

This PR adds:
- Special cmsParagraphs, cmsImages and images
- seeding config for these
- New somponents to display special content
- A fallback image in /public


JohanHjelsethStorstad:
I will suggest people try to look around the whole project before merging, as the PR changes infrastructure that affects every single page.

JohanHjelsethStorstad:
Earlier there were many more images that were "special", but since you can change the image on frontpage (Cms_image) I think it makes no sense to have kappemann be special. I will at a late  stage give CmsImages visibility (or permissions), so a csImage can't point to any image. I guess It does not make sense to have logos as special either, bu in a way I sort of feels that it does make sense, I anyway need a default committee logo in the next PR, so we can look at logos then

PR #203, opened 2024-03-12
Perf/optimize docker

Paulijuz:
I got tiered of waiting 40+ seconds everytime i wanted to launch pn (I know, I have a slow computer) so i looked over the docker setup.

I noticed that the `.dockerignore` file was missing quite a files. When `copy . .` is run it copies over, among other thing, `.git` and `docs`. This adds unecessary size to the image, but more importantly invalidates the docker cache for the base image every time a new commit is created...

Another thing I have done is remove the `src` folder from the base image. I have done this because for the dev image it is binded so there is no need to include the entire project if it is not going to be used anyway. This, for me at least, has reduced the image build time quite a bit as the base image cache is valid for longer. For the prod image i copy the `src` folder over explicitly.

Actually, there are quite many thing in the `.dockerignore` file now and so few that we actually need to copy over so i changed the `Dockerfile` file to explicitly copy over the files we wish to use. This, for me at least, feels more robust as we avoid situations where we accidentaly copy over files and folder which are not used and only take up space and time.

I also looked over the compose file. I reduced the db healtcheck polling period from 3 seconds to 0.5 seconds as there aren't really any downsides to it. I also added proper depends_on conditions for the different services. I changed the pn container to wait for migrations to finnish as I sometimes have experienced crashes when both migrations and pn are trying to access the db at the same time. I also changed the pn container to wait for the db to be ready.

Finally, the biggest improvment. I added the `.next` as a volume for the pn dev container. This folder, among other things, caches the built code during dev. Before everything had to be rebuilt when launching the container, now it can reuse the cache from previously. This has, for me at least, reduced the compile time for the home page from on average above 40 s to around 5 s to 10 s.

Anyway, feel free to disagree. I haven't used docker a lot so I might have done something stupid.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,19 +1,36 @@
+# node
 node_modules
 npm-debug.log
+
+# eslint
+.eslintignore
+.eslintrc.json
+
+# next js
 build
 .next
 
-#you should be explicit in docker-compose about what env-vars should be loaded into a container
+# configs
+# you should be explicit in docker-compose about what env-vars should be loaded into a container
 .env 
 default.env
 next-env.default.d.ts
-
-README.md
-.gitignore
-docker-compose.yml
-Dockerfile
-docker-compose.dev.yml
 nginx.conf
+
+# info
+README.md
+docs
+
+# git
+.git
+.gitignore
+.github
+
+# docker
+.dockerignore
+docker-compose.yml
+docker-compose.dev.yml
+Dockerfile
 
 # prisma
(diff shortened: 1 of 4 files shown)


JohanHjelsethStorstad (approved):
Since it does not affect prod it is fine. I have never had any issues, as docker seams to quite aggressively caches a lot. Often I have had that docker taks 30 secs to start initially in dev, but after that it only takes seconds. But I think this makes sense.

PR #208, opened 2024-03-15
Feat/use editing

JohanHjelsethStorstad:
Note that this PR branches from refactor/server-folder #206
Till now the edit icon just showed if you were Harambe104, now there is a system in place to determine if it should show or not.
This PR does the following:
- requiredPermissions in getUser is now a matrix so one can have and and or statements (it used to be just and), I made this change to mirror the upcoming visibility. Then we will also have a groupVisibilityMatrix.
- Made useUser work more like getUser
- Renamed the files for getUser and useUser and removed the tsx ext. on the latter.
- editModeContext nox has functionality to keep track of all editable content mounted in the dom.
- A new useEditing hook that takes a permissionMatrix (soon also a visibilityMatrix) and uses useUser to determine if the user can edit the content, it returns true if editMode is on and the user can edit the content. 
- Changes all useContext(EditModeContext) to useEditing.


JohanHjelsethStorstad:
Will be replaced by new PR, containing groups and server refactor

PR #209, opened 2024-03-17
Refactor/server folder and error

JohanHjelsethStorstad:
Another day another refactor.
This PR builds on #206 and adds ServerError class for the server folder and a way to translate a function that might throw serverError or return data to a ActionReturn with ```safeServerCall```

Still I have done nothing with groups, and the zoo parse is still not split. I seguest this be refactored so actions have bare-bones  schema only validatiog to right ts types. while /server has the more elaborate schemas with .max .email, ... . The /server folder should throw ServerError type 'BAD PARAMETERS' if the parse goes wrong (with the right path), this will then be picked up in ```safeServerCall``` and converted to ActionReturn


JohanHjelsethStorstad:
See #213

PR #213, opened 2024-03-21
Feat/groups-/server-validation-useEditing

JohanHjelsethStorstad:
A big PR that contains
 - The group system
 - A refactor to create a separate /server directory #209 
 - A new way to write zod validations
 - The useEditing hook #208


Paulijuz (approved):
This is a heafty pr. I've tried to test as much as possible and have fixed all the bugs i came accross. 

There are still some features and kinks to work out. For example:
- We need a better way to add groups to permission roles (right now you have to type in the group id)
- The get user and use user functions should be able to check for group memberships
- The auth folder needs a refactor, there are still a few things left in it that should be moved to the server folder

However, I think this PR is big enough as it is and right now we cannot really move forward without mergeing it as future additions to the project depend upon this refactor. So this PR is greenlit from me.

Paulijuz:
I also think @theodorklauritzen should maybe have a look to confirm that nothing feide related has been broken

JohanHjelsethStorstad:
@theodorklauritzen  can merge it

theodorklauritzen:
Things look good, but there a a few small bugs, but they are not critical.
- Delete collection button is shown on special collections
- Create user from the admin dashboard does not work

I have only tested the UI where there have been changes to the code, but it seems nice :))

theodorklauritzen (approved):

Paulijuz:
Woho! Finally

PR #216, opened 2024-03-24
Refactor/auth get user and use user

Paulijuz:
This PR splits off the memberships and permssions from the user object to their own properties of the session object. The user object no longer has permissions or memberships as properties. This means that `getUser` and `useUser` returns permissions and memberships on the same level as the `user` object. The new usage looks like this:

```ts
 const { user, permissions, memberships, authorized, status} = useUser({
        requiredPermissions,
        shouldRedirect: false,
})
```

Permissions and memberships are always returned as arrays, even if there is no user. If there is no user the permissions array will include the default permissions while the memebrships array will be empty. This is so that one can always safely do `permissions.includes()`.


src/auth/authoptions.ts
```diff
@@ -148,7 +150,7 @@ export const authOptions: AuthOptions = {
             const userId = user ? Number(user.id) : token?.user.id
 
             // TODO - refactor when read user action exists
-            const userInfo = await prisma.user.findUnique({
+            const userInfo = await prisma.user.findUniqueOrThrow({
```
JohanHjelsethStorstad, line 153: This should prob. be inside prismaCall. We need to go through this at some point

src/typings/next-auth.d.ts
```diff
@@ -28,7 +30,9 @@ declare module 'next-auth/adapters' {
 
 declare module 'next-auth/jwt' {
     interface JWT {
-        user: ExpandedUser,
+        user: UserFiltered,
```
JohanHjelsethStorstad, line 33: I guess it makes sense to also split it. in the JWT

JohanHjelsethStorstad (approved):

theodorklauritzen (approved):
LGTM :)))

PR #240, opened 2024-05-01
Feat/notifications

theodorklauritzen:
Fix issues:
- #162 
- #119 
- #15

Issue #139, opened 2024-02-01
feat/skapres


Paulijuz:
The new veven needs to implement the locker reservation system ("skapres") from the old veven.

There is a related issue (#41) to add qr-codes to the lockers so one can reserve them easier than with just the website, but this an improvment on the old system so it isn't necessary to complete this issue.

Issue #41, opened 2023-11-20
feat/qr-code on skapres


theodorklauritzen:
QR-codes should be implemented to the locker reservation system. The each qr code should be uniqe an placed on their corresponding locker. Then, to reserve a specific locker a user could simply scan the QR-code and be taken to the correct reservation page.

PR #245, opened 2024-05-13
Feat/registration

### LINKED ISSUE #10: feat/Fast soelle to member prosess
Should we have a system to make it easy to change status to member right away at Phaesum to avoid a long process of editing each profile later on.

theodorklauritzen:
- Add more registration fields (alergy, phonenumber)
- Add forgot password
- Add verify email
- Add a redirect to /register if the user is not registered, and foes to a page with getUser shouldRedirect is set to true
- Add OmegaId (important milestone for #10)
- Added a way to register soelles to members (Fixes #10)
- Added Qr code libraries ( #139 #41)
- Add UI to admin studyProgrammes
- Automaticly add users to OmegaMembershipsGroups

Caution! Merge the feat/notifications #240 branch and the feat/auth folder #228 first!


src/server/groups/omegaMembershipGroups/read.ts
```diff
@@ -1,15 +1,58 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
     }))
+
+    for (let i = 0; i < results.memberships.length; i++) {
+        const g = results.memberships[i]
```
Paulijuz, line 50: wouldn't a more efficient approach be to get the ids of the omega membership groups

`const groupIds = prisma.omegaMembershipGroup.findMany({ select: { id: true } })`

and then check if there exists a membership between the user and any of the groups?

`prisma.memberships.findUnique({ where: { userId, groupId: { OR: groupIds } } })`

Paulijuz (changes requested):
This PR rickrolled me, therefore I cannot approve it. >:(

Also, I added some comments throughout the changes and is some general feedback:
- Are you supposed to be able to enter any domain name you want into a mail alias. Would it be better to simply enter the name that should be before the @?
- Should the admissions be a database table? This doesn't seem like something that is likely to change in a loooong time so I don't see the benefit of allowing it to be dynamic. Maybe a simple enum would suffice? Then you could reduce the complexity quite a bit.
- I'm not sure if this is the PR that did it, but the create new user form is broken.
- A lot of the forms revert to the old values after submission, but I'm guessing a lot of this will be fixed when we do a second pass over the UI.
- I think turning on a lot of notification channels is very tedious. Especially since you have to press the submit button for each channel.

I don't know if I was too picky (a lot of it was just grammar) or not so fix the issues you deem worthwhile to fix and then I'll approve this PR. All inn all this PR brings a lot of critical features. It's really nice to se some good progress.

theodorklauritzen:
Thanks for the feedback @Paulijuz 

> Are you supposed to be able to enter any domain name you want into a mail alias. Would it be better to simply enter the name that should be before the @?

Some users should be able to add different domains than the current domain. Maybe we want the alias POSTMASTER@sanctus.omega.ntnu.no for example. But I agree that it's unnecessary in 99% of cases not to add this automatically and prevent users from typing the wrong domain.

> Should the admissions be a database table? This doesn't seem like something that is likely to change in a loooong time so I don't see the benefit of allowing it to be dynamic. Maybe a simple enum would suffice? Then you could reduce the complexity quite a bit.

I had an idea to use this more generally, like admissions to DGR. But after some thought I don't think this will be used, and then it's just extra unnecessary complexity.

> I'm not sure if this is the PR that did it, but the create new user form is broken.

I have some thoughts about this. I think the user form should take in the fields `username`, `firstname`, `lastname` and `email`. Then when you create a user, they should get an email with a link that leads to the registrations page. And there they can create a password add a mobile number etc. Do you have any thoughts on this as well @JohanHjelsethStorstad ?

> I think turning on a lot of notification channels is very tedious. Especially since you have to press the submit button for each channel.

I totally agree the UI should be a lot better. I was not sure how the profile page design would be. But the best is probably to put the settings on a different page.

Paulijuz:
> Some users should be able to add different domains than the current domain. Maybe we want the alias [POSTMASTER@sanctus.omega.ntnu.no](mailto:POSTMASTER@sanctus.omega.ntnu.no) for example. But I agree that 99% of cases we just use omega.ntnu.no.

Then I understand your design choice. Ideally, I still think there should be something in the UI to indicate which domains are valid tho. Currently you either have to know which domains are valid or look through the list of used domains. Maybe you could have a drop down of all the valid domains? Though this is not really important, this page will be used by very few people (only admins) so I think how it is now is good enough. We can always improve later on.

JohanHjelsethStorstad:
I agree that we should have a fixed number of domains allowed

JohanHjelsethStorstad:
As for the createNewUser form it was meant as an admin-form. So it should probably  take as little info as possible then make the new user give the rest

src/server/mail/read.ts
```diff
@@ -0,0 +1,484 @@
...
+ * @returns A Promise that resolves to a MailFlowObject.
+ */
+async function readAliasTraversal(id: number): Promise<MailFlowObject> {
+    const results = await prismaCall(() => prisma.mailAlias.findUniqueOrThrow({
+        where: {
+            id,
+        },
+        include: {
+            mailingLists: {
+                include: {
```
JohanHjelsethStorstad, line 78: This should have some includers. Often (look at cms modules) when we had long includes I started aking them in ConfigVars.

src/auth/jwt.ts
```diff
@@ -1,3 +1,10 @@
+import 'server-only'
+import { JWT_ISSUER } from './ConfigVars'
+import { ServerError } from '@/server/error'
+import jwt, { JsonWebTokenError, TokenExpiredError } from 'jsonwebtoken'
+import type { OmegaJWTAudience } from './Types'
+
+
 // See https://www.rfc-editor.org/rfc/rfc7519#section-4.1 for the
 // JWT payload specification.
 export type JWT<T = Record<string, unknown>> = T & {
```
JohanHjelsethStorstad, line 10: As I have said I think I will make a directory for jwt

src/app/components/User/CreateUserForm.tsx
```diff
@@ -18,16 +18,15 @@ export default function CreateUserForm({ className }: PropTypes) {
...
                 successCallback={refresh}
             >
-                <TextInput label="username" name="username" key={uuid()}/>
-                <TextInput label="password" name="password" key={uuid()}/>
-                <TextInput label="confirm password" name="confirmPassword" key={uuid()}/>
                 <TextInput label="email" name="email" key={uuid()}/>
+                <TextInput label="username" name="username" key={uuid()}/>
                 <TextInput label="first name" name="firstname" key={uuid()}/>
                 <TextInput label="last name" name="lastname" key={uuid()}/>
+                <p>Når en bruker lages vil brukeren få tilsendt en epost, men en link for å fullføre registreringen.</p>
```
JohanHjelsethStorstad, line 29: med en link

src/app/components/OmegaId/identification/OmegaId.tsx
```diff
@@ -0,0 +1,15 @@
+'use server'
+
+import OmegaIdElement from './OmegaIdElement'
```
JohanHjelsethStorstad, line 3: I think all of this needs a generalization to use qr-codes for skapres, registrations, ...

JohanHjelsethStorstad: Just remember that it is written I guess

src/actions/news/update.ts
```diff
@@ -26,6 +27,12 @@ export async function publishNewsAction(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     shouldPublish: boolean
 ): Promise<ActionReturn<Omit<SimpleNewsArticle, 'coverImage'>>> {
+    dispatchSpecialNotification(
```
JohanHjelsethStorstad, line 30: A little unsure if this should be done in the action or in the function??

src/actions/omegaquotes/create.ts
```diff
@@ -21,5 +22,11 @@ export async function createQuoteAction(
     if (!parse.success) return createZodActionError(parse)
     const data = parse.data
 
-    return await safeServerCall(() => createQuote(user.id, data))
+    const results = await safeServerCall(() => createQuote(user.id, data))
+
+    if (results.success) {
```
JohanHjelsethStorstad, line 27: Same as for news. We just need a convention.

JohanHjelsethStorstad (commented):
Looks good

JohanHjelsethStorstad (changes requested):
The above

src/server/omegaid/generate.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const payload: OmegaIdJWT = {
+        sub: data.id,
+        usrnm: data.username,
+        gn: data.firstname,
+        sn: data.lastname,
+    }
+
+    const token = jwt.sign(payload, privKey, {
```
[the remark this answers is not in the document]
theodorklauritzen: GenerateJWT uses symmetric keys, this function uses asymmetirc keys. The function should be moved to make this clear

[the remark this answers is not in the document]
JohanHjelsethStorstad: Then I think you should have two methods in the jwt file

src/app/components/OmegaId/identification/OmegaId.tsx
theodorklauritzen: The next-qrcode library has a simple api. I don't think we need to make an abstraction on top of that. See `OmegaIdElement.tsx` for an example.

src/server/mail/read.ts
theodorklauritzen: I can move the include statements to a ConfigVars file, the the include statements will probably only be used one place.

src/server/auth/resetPassword.ts
```diff
@@ -0,0 +1,49 @@
+import 'server-only'
+import { resetPasswordValidation } from './validation'
+import { ServerError } from '@/server/error'
+import { readUser } from '@/server/users/read'
+import { verifyJWT } from '@/auth/jwt'
+import { sendResetPasswordMail } from '@/server/notifications/email/systemMail/resetPassword'
+
+export async function resetPasswordByEmail(email: string): Promise<string> {
```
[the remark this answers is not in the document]
theodorklauritzen: Hmm, I have combined the functions. I think it is nice to have the functions sending email in the same folder, but I'm unsure why, so we can change that at place the function from where they were called from.

src/styles/_mixins.scss
```diff
@@ -112,6 +112,18 @@
     }
 }
 
+@mixin screenMobile {
```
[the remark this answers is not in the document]
theodorklauritzen: I agree, and it would be nice to discuss how we should implement different screen sizes. They should maybe be called sm, lg instead of mobile as I have written.

theodorklauritzen:
I have fixed most of the feedback you gave. I have also created a JWT folder, with all the JWT functionality.  The things I haven't fixed yet are:

- mobile mixin on all pages
- Includer for long queries (I don't feel like the queries should be in config since they are not config)
- The user form is still broken, but it would be nice to discuss how this should function. (And I is fixed in a later PR)

src/server/notifications/email/mailHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        await this.handleNewMail()
+    }
+}
+
+function mailHandlerSingleton() {
+    return new MailHandler()
+}
+
+declare const globalThis: {
```
Paulijuz, line 136: this should be moved to src/typings/global.d.ts

src/server/notifications/email/systemMail/userInvitivation.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { UserInvitationTemplate } from '@/server/notifications/email/templates/userInvitation'
+
+
+export async function sendUserInvitationEmail(user: UserFiltered) {
+    const jwt = generateJWT('verifyemail', {
+        sub: user.id,
+        email: user.email,
+    }, userInvitationExpiration)
+
+    const link = `${process.env.NEXTAUTH_URL}/register?token=${jwt}`
```
Paulijuz, line 15: It feels a bit iffy using an environment variable created for next auth elswhere in our project. Especially since you've used `DOMAIN` elsewhere, for example in `dispatch.ts`. Logically we should be using `DOMAIN` and set it to https://localhost:80. Then we could set `NEXTAUTH_URL=DOMAIN` to simplify things. However, I guess it might look a bit weird having localhost plastered all over the website instead of omega.ntnu.no, but that might not be a bad thing because then we could easily spot hard coded URLs. Another option would be to add another environment variable so we could differentiate between the actual link url and the domain which should appear for visual thing only.

src/server/users/update.ts
```diff
@@ -17,6 +29,59 @@ export async function updateUser(id: number, rawdata: UpdateUserTypes['Detailed'
...
+            }
+        }))
+
+        return {
+            verified: true,
+            email,
+        }
+    }
+
+    if (storedUser.feideAccount?.email !== email) {
```
Paulijuz, line 68: The condition in this if statement will always be true due to the previous if statement. So it's unnecessary.

src/server/notifications/email/mailHandler.ts
Paulijuz: you can see how it's done with prisma :)

Paulijuz (dismissed):
Nice work! Really like the new notification settings menu and the comment about how mail flow works is very helpful. I also like the new login color scheme, although we will need a pass over the UI to make all the pages consistent with each other. 

I haven't done as much extensive testing since I'm guessing the core functionality works the same as before. I still managed to find some UI bug tho, all of them are _minor_ things related to the new notification settings page.
- If you make further changes by toggling channels after having pressed save, the save button will still disappear. 
- If I enable all the channels with the top check button I _sometimes_ get the error message "The methods must a subset of the available methods". I'm guessing this is because the top check button checks the disabled check buttons somehow.

I also made some small comment. Keep up the good work!

theodorklauritzen:
The latest comments from the review from @Paulijuz  should have been resolved by now. Hovever I have not bee enable to reproduce the bug:

> If I enable all the channels with the top check button I sometimes get the error message "The methods must a subset of the available methods". I'm guessing this is because the top check button checks the disabled check buttons somehow.

Therefore it's probably not fixed, but I have changed how the page works and that can be why I'm unable to reproduce the bug.

@JohanHjelsethStorstad do you want to give me another review?

.eslintignore
```diff
@@ -2,4 +2,5 @@ node_modules/
 .next/
 dist/
 src/prisma/prismaservice/generated/
-src/prisma/prismaservice/dist/
\ No newline at end of file
+src/prisma/prismaservice/dist/
+src/typings/global.d.ts
```
johanMedistim, line 6: O if you look in global.d.ts I used disable no var for one line. Either remove this or ignore the file. Now the ignore is pointless

johanMedistim, line 6: f

src/server/notifications/channel/validation.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function validateNewParent(channelId: number, newParentId: number, channels: ExpandedNotificationChannel[]): boolean {
+    return findValidParents(channelId, channels).map(c => c.id).includes(newParentId)
+}
+
+export function findValidParents(channelId: number, channels: ExpandedNotificationChannel[]): ExpandedNotificationChannel[] {
+    const validParents = channels.filter(c => c.id !== channelId)
+    // Remove chrildren of the current channel
+    const channelIDS = new Set(validParents.map(c => c.id))
+    while (true) {
```
johanMedistim, line 53: This seams a little dangerous

JohanHjelsethStorstad, line 53: This feels dangerous

src/typings/global.d.ts
```diff
@@ -1,7 +1,9 @@
+import type { mailHandlerSingleton } from '@/server/notifications/email/mailHandler'
 import type { Prisma, PrismaClient } from '@prisma/client'
 import type { DefaultArgs } from '@prisma/client/runtime/library'
 
 declare global {
    // eslint-disable-next-line no-var
    var prisma: PrismaClient<Prisma.PrismaClientOptions, never, DefaultArgs>
+   var mailHandler: ReturnType<typeof mailHandlerSingleton>
```
JohanHjelsethStorstad, line 8: Above I have ignored the no-var rule. Either add it to this line too and remove the file from .eslintignore, or remove my no-var ignore, since now it is pointless.

src/server/notifications/Types.ts
```diff
@@ -0,0 +1,26 @@
...
+    email: true,
+    emailWeekly: true,
+    push: true,
+} satisfies Record<NotificationMethods, boolean>
+
+export const allMethodsOff = {
+    email: false,
+    emailWeekly: false,
+    push: false,
+} satisfies Record<NotificationMethods, boolean>
```
JohanHjelsethStorstad, line 26: These are not types?

JohanHjelsethStorstad (approved):
I think it mostly looks fine now

JohanHjelsethStorstad (dismissed):
It looks fine, but there is logic here that is hard to understand. Also we can fix some small things I have commented

theodorklauritzen:
@JohanHjelsethStorstad fixed your comments.

I'll refactor a few things in this PR while working on push notifications.

JohanHjelsethStorstad (approved):

Issue #229, opened 2024-04-09
fix/unable to access 404 pages while not logged in


willijs:
When accessing 404 pages while not logged in, the page crashes because the site tries to access the page in an infinite loop

PR #260, opened 2024-06-06
Feat/jobads

### LINKED ISSUE #149: feat/jobads
Jobads - contact contactor (pun intended) to hear what they want. Do not look to how it is done on the old vev. A add should probably use Article or ArticleSection component

JohanHjelsethStorstad:
This closes #149.
Still we should add more metadata on the adds. But this gives us the same functionality as was on the old veven, just infopage free.


JohanHjelsethStorstad:
Note that this PR changes the way defaultPermissions are handled. Now they are fetched one time server side and stored in a context. Pay therefore attention to the changes in useUser, and make sure it still makes sense.

src/app/jobads/CreateJobAdForm.module.scss
Paulijuz, line 1: Is this file supposed to be empty?

src/app/jobads/[...orderAndName]/page.module.scss
Comment (Paulijuz) [inline src/app/jobads/[...orderAndName]/page.module.scss:1]: Is this file also supposed to be empty?

src/server/jobAds/ConfigVars.ts
```diff
@@ -0,0 +1,19 @@
+import { articleRealtionsIncluder } from '@/cms/articles/ConfigVars'
+
+export const jobAdArticleRealtionsIncluder = {
+    article: {
+        include: articleRealtionsIncluder
+    }
+} as const
```
Paulijuz, line 7: For better intellisense you can add "satisfies Prisma.JobAdInclude" after "as const".

src/app/jobads/CreateJobAdForm.tsx
Paulijuz, line 1: The form should be in Norwegian.

Paulijuz (approved):
Great job(ads)! ;)

I found only one major issue. The permissions for jobads are a bit broken. When I disabled all the permissions for jobads and tried to go to the joabds page it threw and error. When I only disabled the create, read and update permissions the create jobad button was still visible (this is only a visual thing) and I could edit the article part of the jobad (i.e. create, remove and edit sections, and edit the title, but not the company or delete the jobad).

I also made some no review comments on some really minor things.

I'll approve this PR as the permission problems can be fixed at a later date and they're not too big an issue for now. If you want to you can fix the issues and I'll approve again. If not I can create a new issue for the jobad permission.

src/app/jobads/currentJobAds.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @param not - pass it not: a id of a jobad to exclude from the list
+ */
+export default async function CurrentJobAds({ not }: PropTypes) {
+    const res = await readJobAdsCurrentAction()
+    if (!res.success) {
+        throw res.error ?
+            new Error(res.error[0].message) :
+            new Error('unknown error reading jobad')
```
Paulijuz, line 16: I think you should check for if the error code is unauthorized or unauthenticated. Then you could show the 404 page (or an unauthorized page if we get that in the future).

Another option would be to check if the user has the correct permissions at the beginning, but then as you have pointed out earlier we would be checking the users permissions twice. Another issue with this is that if the action permissions were to change you would also have to change the permissions in the component.

Maybe, since this is a common occurrence, we could write a monad similar to useActionCall that would check the error code beforehand and if it's unauthorized or unauthenticated it would do the appropriate action.

JohanHjelsethStorstad: We have talked about this before. If (!res.success) we should have a standard unwrapper function for actionerrors

JohanHjelsethStorstad:
The thing with updating the article is not a problem. This must be authed i the article component so it is visibility. Jobads will be a special visibilitet purpose

src/app/jobads/currentJobAds.tsx
Paulijuz: Forgot about that, but I agree.

Paulijuz:
Ah ok, so it will be fixed when visibility is done.

src/app/jobads/CreateJobAdForm.module.scss
JohanHjelsethStorstad: Things will be added when metadatais added got jobads. We probably need a sidebar

Paulijuz (approved):
No logic changes in the code since my last review so it looks good to me!

I also made this PR resolve issue #229 as the new default permission context fixes the issue.

src/app/jobads/CreateJobAdForm.tsx
Paulijuz: we'll fix this in the future

PR #261, opened 2024-06-06
chore(deps): clean up dependencies

JohanHjelsethStorstad:
This PR:
- Upgrades react-pdf
- removes unused typing dependency
- upgrades eslint-import-resolver-typescript and eslint-plugin-import

I was not able to upgrade eslint to 9.0.0 as eslint-plugin-import does not support it I think


theodorklauritzen:
It seems like the upgrade of react-pdf breaks the ombul page

JohanHjelsethStorstad:
Hah. I tested it in safari and it worked but on chrome it is broken

Paulijuz:
I think the building issue is a problem with Next js. According to [this comment](https://github.com/wojtekmaj/react-pdf/issues/1822#issuecomment-2172868407) a fix is waiting for the next release. In the mean time setting `swcMinify: false` in `next.config.js` seems to fix the problem.

JohanHjelsethStorstad:
I don't think changing the next Is a great idea. Have used too much time on this already, so I will revert react-pdf I think, and just hope it get  better in the near future

JohanHjelsethStorstad:
Or just wait to merge this

Paulijuz:
I agree that playing around with the next settings isn't a good idea. We can wait until the next update (which hopefully will fix the issue).

Paulijuz:
I'll close this pr until then

PR #262, opened 2024-06-08
Feat/postfix & nginx

theodorklauritzen:
Change of the production evironment to include, postfix, nginx with SSL, and replicas of the projectnext image. There is probably still a bug with the mail forwarding with Postfix, due to the PR #245 not being merged yet.


Paulijuz (approved):
Making a monorepo was a good idea. I think this is very clean. The e-mail query looks also quite clean and readable. Nice work!

docker-compose.dev.yml
```diff
@@ -1,96 +1,100 @@
...
-    depends_on:
-      - projectnext
-    links:
-      - projectnext
+    nginx:
+        image: nginx:1.21.3-alpine
+        ports:
+            - 80:80
+        volumes:
+            - nginx/nginx.dev.conf:/etc/nginx/nginx.conf
```
Paulijuz, line 87: Just noticed a './' at the beginning s missing here. This makes docker think that this is a volume instead of a bind. Whoopsee...

PR #263, opened 2024-06-09
Docs/submodule

Paulijuz:
The github wiki page requires the documentation files to be in its own repo (don't ask me why). Therefore documentation now lives at the repo "projectNext.wiki". To make things a bit easier however I have added the new repo as a submodule. It's located in the same folder as the doc files were in previously.

To clone the submodule (and update it in the future) you can run: `git submodule update --init`

If you're cloning the repo from scratch you can run: `git clone --recurse-submodules <link to repo>`

When you cd into the docs folder now it will act as it's own repo with its own git history.


theodorklauritzen (approved):
Great idea with the wiki, it will be more intuitive than a folder for later generations to use.

theodorklauritzen:
Hmm, @JohanHjelsethStorstad should maybe have a look at the PR before I merged it, sorry for the late notice.

PR #269, opened 2024-06-17
Refactor/split prisma schema

Paulijuz:
FINALLY!

I split the schema file into multiple files based on the categories we already had. I don't know if this is the best way as some of the categories only have one model (for example "omegaOrder") so feedback is greatly appreciated. 

I also had to move the schema from omegaweb down into its own folder as it was confusing the vs code extension when editing the schema for pn. The reason is that:
> [...] this is an artifact of how the VS Code extension detects the parent or main schema folder. It goes up until it does not find a .prisma file anymore. The CLI and other tooling gets an explicit Prisma schema path via --schema or some other configuration, so it knows that main folder out of the box. The VS Code extension unfortunately does not, which is why we need that logic.

Source: https://github.com/prisma/prisma/discussions/24413#discussioncomment-9675798


theodorklauritzen (approved):
This looks fantastic! It is much easier to work with smaller files than one large one! As you say some files only contain one model, which is quite little, but I don't have a better convention than to separate the files as you have done.

JohanHjelsethStorstad (approved):
Loos good. I think the current splitting is good

PR #278, opened 2024-06-26
Chore/cursor based pagination

JohanHjelsethStorstad:
I have managed to make the paging system even more complicated. This PR changes offset based pagination to cursor-based pagination. This should be more optimal for larger tables. I have also changed some structure to go along with this, like the addition of a server/pagination directory.


Paulijuz:
Awesome! From what I know and what an old vever told me cursor based pagination is much preferred over skip+take pagination so this is very nice. I'll try to do a full review tomorrow, but from a quick looks it doesn't seem too complicated.

Paulijuz (approved):
Looks good to me! The change to cursor based didn't make the code that much more complicated for the added benefit of improved performance. It's also nice to see more comments. <3

The only thing I want to comment on is that with cursor based pagination one can't easily go to a specific page. However, since we're primarily using endless scrolling this isn't a problem, but it's something that we might need to reimplement (not that it's hard) if we ever need the ability to jump to a specific page.

theodorklauritzen (approved):
Looks good to me

JohanHjelsethStorstad:
> Looks good to me! The change to cursor based didn't make the code that much more complicated for the added benefit of improved performance. It's also nice to see more comments. <3
> 
> The only thing I want to comment on is that with cursor based pagination one can't easily go to a specific page. However, since we're primarily using endless scrolling this isn't a problem, but it's something that we might need to reimplement (not that it's hard) if we ever need the ability to jump to a specific page.

How can it not?? My thought was just to use a query param that sets the state of the last cursor position in the url for pages

PR #282, opened 2024-07-08
Feat/profile page

AndreasVJ:
Improved graphics for the profile page. User settings page still needs improvements


src/app/users/[username]/page.module.scss
```diff
@@ -0,0 +1,174 @@
...
+
+.silver {
+    background: linear-gradient(to bottom right, #ededed, #bdbdbd);
+}
+
+.gold {
+    background: linear-gradient(to bottom right, #B69312, #FFDF6F);
+}
+
+.diamond {
```
Comment (Paulijuz) [inline src/app/users/[username]/page.module.scss:26]: I might be blind, but are these classes currently used somewhere? I'm guessing they're for the "kappemann" flairs?

src/app/users/[username]/page.tsx
```diff
@@ -11,40 +17,104 @@ type PropTypes = {
...
+    const profile = await readUserProfile({username})
+
+    if (!profile) {
+        notFound()
+    }
+
+    const groupIds = profile.memberships.map(group => group.groupId)
+    const committees = await readCommitteesFromIds(groupIds)
+
+    const studyProgramme = await prismaCall(() => prisma.studyProgramme.findFirst({
```
Comment (Paulijuz) [inline src/app/users/[username]/page.tsx:38]: This should be an action. If you think we can refactor this at a later date add a todo comment.

src/app/users/[username]/page.tsx
```diff
@@ -11,40 +17,104 @@ type PropTypes = {
...
+                    <Image className={styles.profilePicture} image={profileImage} width={240}/>
+                </div>
+                <div className={styles.header}>
+                    <h1>{`${profile.firstname} ${profile.lastname}`}</h1>
+                    <p className={styles.studyProgramme}>{studyProgramme.name} {`(${studyProgramme.code})`}</p>
+                    <div className={styles.committeesWrapper}>
+                        {committees.map(committee => <div className={styles.committee}><p>{committee.name}</p></div>)} {/* TODO change to your own committee title instead of committee name*/}
+                    </div>
+                    <hr/>
+                    <p className={styles.orderText}>{profile.sex == "FEMALE" ? "Syster" : "Broder"} uudaf {order}´dis orden i Sanctus Omega Broderskab</p>
```
Comment (Paulijuz) [inline src/app/users/[username]/page.tsx:87]: Always use tripple equals in javascript. I think this is currently one of the linter errors for this branch. Also, I think a better approach would be to use a mapping something like this:

```ts
const MEMBERSHIP_RELATION = {
    FEMALE: 'Syster',
    MALE: 'Broder',
    OTHER: 'Soesken',
} satisfies Record<SEX, string> // Ensures that all possible SEX values exists as keys in the object
```

So that you could then do:
```ts
{MEMBERSHIP_RELATION[profile.sex ?? 'OTHER']} uudaf {order}´dis orden i Sanctus Omega Broderskab
```

src/app/users/[username]/page.module.scss
Comment (Paulijuz) [reply src/app/users/[username]/page.module.scss:26]: Never mind, saw your comment further down.

Paulijuz (changes requested):
The profile page looks awesome now! 

However, this PR cannot be merged until the linter has been satisfied. You can run `npm run lint -- --fix` to fix the issues that can be fixed automatically. The remaining ones you'll hav to (sadly) fix manually. If you run `npm run lint -- --quiet` you will only see the linting _errors_ (the ones preventing the PR from being merged).

I also added a few comments.

Keep up the good. I love  that the harambe tradition lives on.

src/prisma/schema/user.prisma
```diff
@@ -10,10 +10,13 @@ model User {
   email         String        @unique
   firstname     String        @default("[Fjernet]")
   lastname      String        @default("[Fjernet]")
+  bio           String        @default("")
```
JohanHjelsethStorstad, line 13: cms paragraph?

src/app/users/[username]/page.tsx
```diff
@@ -11,40 +17,104 @@ type PropTypes = {
...
-    })
+    }))
 
-    if (!userProfile) {
-        notFound()
+    if (!studyProgramme) {
+        throw new Error("studyProgramme not found")
+    }
+
+    const order = studyProgramme.group.memberships[0].omegaOrder.order
```
Comment (JohanHjelsethStorstad) [inline src/app/users/[username]/page.tsx:65]: There will come functionality to read users with more info atached I think

Comment (JohanHjelsethStorstad) [reply src/app/users/[username]/page.tsx:65]: Like study programme, year etc...

src/styles/_mixins.scss
```diff
@@ -23,7 +23,6 @@
     margin: variables.$gap;
     color: colors.$black;
     border: none;
-    width: 100%;
```
JohanHjelsethStorstad, line 26: This is a little dangerous to change since it potenitially changes a lot of UI. Have somone looked through it all?

src/app/users/[username]/settings/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (!profile) {
+        return notFound()
+    }
+
+    const permissions = await readPermissionsOfUser(profile.id)
+    const memberships = await readMembershipsOfUser(profile.id)
+
+    return (
+        <div>
+            <Link href={`/users/${username}`}>Tilbake</Link>
```
Comment (JohanHjelsethStorstad) [inline src/app/users/[username]/settings/page.tsx:39]: We might potentially want a layout file for the profile part

src/app/users/[username]/page.tsx
```diff
@@ -11,40 +17,104 @@ type PropTypes = {
...
+
+    let profileImage = profile.image
```
Comment (JohanHjelsethStorstad) [inline src/app/users/[username]/page.tsx:67]: use const as mutch as possible

Comment (JohanHjelsethStorstad) [reply src/app/users/[username]/page.tsx:67]: prefer things like profile.image ? profile.image : await readSpecialImage(...)

JohanHjelsethStorstad (changes requested):
I think we need to think a little about how we do profile images. We might do it with CmsImage under the hood (this is the way committee logos do it). But since non of the profileImage logic is implemented it is fine for now. Having.a ProfileImage component might be just as good, defaulting to special image: DEFAULT_PROFILE_IMAGE. It is a very nice start, and it is usful me to get this merged now so I can test visibility.

src/app/users/[username]/page.tsx
Comment (JohanHjelsethStorstad) [reply src/app/users/[username]/page.tsx]: Changed my mind. I put it in users/ConfigVars

src/styles/_mixins.scss
JohanHjelsethStorstad: I have solved the one place I found but there might be more. But I think it is more logical now

JohanHjelsethStorstad:
So I fixed the listing and did some refactoring. I also changed some of the logic to do with authing. The part in users/[username]/page where the study programmes and so on are fetched will be refactored in feat/visibility. On that branch I have logic to infer names of groups and so on. My plan is to work on the setting screen from there. As I need it to debug visibility

src/app/users/[username]/page.tsx
Comment (JohanHjelsethStorstad) [reply src/app/users/[username]/page.tsx:51]: As I mentioned I will change this on the visibility branch. There I have made functions to infer information about users and groups. The way I infer a users order is by looking at their membership order to the omega membership group, not study program. I think this makes more sense as if you do not have a membership to the omega member group you do not even have a order and your profile should say soelle or external

JohanHjelsethStorstad (approved):

theodorklauritzen:
@AndreasVJ Wow, the profile page looks amazing!😍

PR #302, opened 2024-08-22
Refactor/folder names

Paulijuz:
This PR simply renames three folder.

`server` has become `services` as after viewing several next js projects it seems that this is the most common name for this type of design pattern. A so called "service layer".

`components` has become `_components` so that it (1) is easier to find (it'll be at the top of the `app` folder now), (2) next js will not treat it as a route (not that that'll ever be a problem since there never should be a file named page.tsx in there anyways), and (3) it's clearer that it's not a route since its name is different from everything else in that folder.

`context` has become `contexts`. This is just because it annoyed me that _all_ the other folders in `src` except this were plural.

Feel free to disagree, but these are just folder names so shouldn't matter too much. \:)


JohanHjelsethStorstad (approved):
I think this looks fine. _components is a good idea. contexts is also a nice (minor change). I don't think that the renting of server matters

JohanHjelsethStorstad (approved):

PR #306, opened 2024-08-28
Chore/api infrastructure

JohanHjelsethStorstad:
This PR adds basic functionality for apis
- Most importantly api keys can be created dynamically
- They are stored hashed + encrypted (just like passwords) in the database
- An api key can have permissions
- We now have a nice frontend to CRUD api keys

This PR also changes
- Makes Permission config more verbose
- Adds components for displaying Permissions
- Changes the permission admin pages to use this
- Also adds new functionality to <Form /> to automatically navigate and/or refresh on success.

Still remains:
- Actually authing based on provided API key, we wait with this since it will use the Auther and ServerSession system to come soon. The idea is that an API key can generate a ServerSession (user (= null), permissions, memberships = []) just like a user through nextAuth
- Actually writing some APIs, this will mostly be a question of just wrapping service-functions


JohanHjelsethStorstad:
I also added a folder for date functionality. I now agree that we need a lib folder

Paulijuz (changes requested):
This looks very nice!

The refactoring of encryption and hashing looks nice. The new permission components also look great!

There are a few things I have commented on. Mostly small things that might not need to be dealt with. Also, there are a few selling errors here and there, but they can be fixed at a later.

When you have fixed/dismissed the comments I'll approve.

JohanHjelsethStorstad:
All fixed @Paulijuz

Paulijuz (approved):
Look good 2 me!

PR #309, opened 2024-08-30
Feat/courses infrastructure schools

JohanHjelsethStorstad:
This PR adds:
- Infrastructure to start work on "fagveven" consisting of courses and schools
- Basic db schema for courses 
- schema for schools
- Backend for schools
- Frontend for schools both admin and regular.
- Some styling generalisations

We might still want to add more feats to schools. Maybe people want to write something about the school they went to exchange to? The main part: the courses is still to be implemented. @mokamv


src/prisma/schema/course.prisma
```diff
@@ -0,0 +1,39 @@
...
+
+  @@id([courseId, studyProgramId])
+}
+
+model CourseApprovedAs {
+  id                 Int    @id @default(autoincrement())
+  courseId           Int
+  courseApprovedAsId Int
+  course             Course @relation(fields: [courseId], references: [id], name: "thisCourseHasBeenApprovedAs")
+  courseApprovedAs   Course @relation(fields: [courseApprovedAsId], references: [id], name: "thisCourseHasBeenUsedToApprove")
```
theodorklauritzen, line 35: Maybe add a relation to a user as well

JohanHjelsethStorstad: I agree. We can either add it now or @mokamv can add It later, I think there are probably more obvious fields missing in this schema

theodorklauritzen:
Looks good. However, I suggest we keep as much admin functionality under `/education` rather than `/admin`. This will make it easier to update content directly from Fagveven without navigating to `/admin`. Additionally, this approach will prevent us from designing specific features to be admin only by design. In the future maybe we want people to be able to add schools. I think we should allow this from the start, however the design makes this process hard for regular users.

JohanHjelsethStorstad:
> Looks good. However, I suggest we keep as much admin functionality under `/education` rather than `/admin`. This will make it easier to update content directly from Fagveven without navigating to `/admin`. Additionally, this approach will prevent us from designing specific features to be admin only by design. In the future maybe we want people to be able to add schools. I think we should allow this from the start, however the design makes this process hard for regular users. The go to admin page helps, but all the admin functionality could be integrated directly.

I sort of agree, but my idea is really that we will need three permissions for schools: READ, CONTRIBUTE, ADMIN. I think we should lock some things. like deletion from people contributing. I thought the admin panel would be mostly for that for stuff like that. We could have ADMIN stuff also on the regular /education page and just hide it for regular contributors. But I don't see the problem in having a more concise admin page too. (we have that for example for users: one can add users from both /users and /admin/users. Here it is done by reusing a generic UserCreate component). If you want to show admin about a school to ADMINs also on the /education/schools page I think this can be done similarly by refactoring the admin page to use components like <SchoolAdmin> and using that both on the /education/schools/[name] (component only shown if editing mode is on) page and /education/schools/[name]. At this point each school does not even have its own page. This should be implemented later.

theodorklauritzen (approved):
Nice, then it looks good. It would be nice with disguising how we should design admin functionality in a meeting soon

JohanHjelsethStorstad:
We should. I generally agree that admin should be near the content (hidden behind Editmode if the admin seriously changes how the content will look so that also admins can see what the content lokka like without all the clutter from admin). But I think if done well, having separat admin pages in addition can give a more concise view of what is going on, without all the clutter of complicated UI.

PR #313, opened 2024-09-07
Chore/api auth and translation service methods

JohanHjelsethStorstad:
This PR adds the Auther and the ServiceMethodHandler and ServiceMethod classes. It still needs more work as the type validation is now unnecessary. I also think the Auther can be made clearer. The idea of the Auther is the typing:

Auther<Static, Dynamic> = Static -> Dynamic -> Session -> AuthResult


src/auth/auther/Auther.ts
```diff
@@ -0,0 +1,88 @@
...
+        session: SessionMaybeUser,
+        dynamicFields: DynamicFields
+    ): CheckReturn<UserRequieredOut>
+
+    public auth({
+        session: sessionIn,
+        dynamicFields
+    }: {
+        session: SessionMaybeUser,
+        dynamicFields: DynamicFields
```
theodorklauritzen, line 51: Why isn't this an optional field? The syntax `dynamicField: undefined` is a bit ugly

JohanHjelsethStorstad: It is very hard to get ts to agree to that. I will look info it when I change it to be just functional

src/services/ServiceTypes.ts
```diff
@@ -0,0 +1,211 @@
...
+    WantsToOpenTransaction extends boolean
+> = PrismaPossibleTransaction<WantsToOpenTransaction> | 'NEW'
+
+
+export type AuthRunConfig = {
+    withAuth: boolean,
+}
+
+type ConfigType<Params> = {
+    params: Params,
```
theodorklauritzen, line 33: I also think this should be optional

src/services/ServiceTypes.ts
```diff
@@ -0,0 +1,211 @@
...
+export type Client<
+    WithValidation extends boolean,
+    DetailedType,
+    Params,
+    Return,
+    WithAuthParam extends boolean,
+    WantsToOpenTransaction extends boolean
+> = {
+    client: (
+        prisma: PrismaTransactionOrNew<WantsToOpenTransaction>
```
theodorklauritzen, line 67: The `'NEW'` keyword works, but is it easier to just accept zero arguments, and then use a new prisma client?

theodorklauritzen:
This looks good after the conversations are answered / fixed :))

src/services/ServiceTypes.ts
JohanHjelsethStorstad: I belive NEW is more clear. It's semantics, I think passing null or nothing is a little implicit.

JohanHjelsethStorstad:
Well I have made it slightly better now I think. (More consistent). I think this pattern is quite good at this point

theodorklauritzen (approved):
Toppers

Paulijuz (commented):
This looks good, but I also wanted to chime in my two cents before this gets approved.

I like the overall pattern, I just have two questions.
1. Do we still need validation after this? Because if server actions and api endpoints will simply call service methods without doing anything more I don't see a reason for a distinction between type and detailed validation. Dropping the validation class and just using basic zod schemaas would simplify the design quite a lot I think.
2. I still don't understand the distinction between service methods and service handlers. Will there be a one to one mapping between service methods and service handlers? If not then when will you call a service method with a different handler? If yes then why the distinction?

JohanHjelsethStorstad:
There are some oddities with how forms submit data. For example a checkbox submits 'on'. We may let the validation In the service worker take in 'on' (or a bool for calls from the api) and convert it to a bool there directly. I also want to say that we should standardise some of the validation I think. Like now people write .max(number, str). This error string should be standard. So I think. in general we will need to rethink validation. We could make some custom validation stuff wrapping zod like val.checkboxOrBool().

There will always be 1:1 between handlers and methods. The handlers are not meant to be called directly, they are only split for code splitting purposes.

PR #320, opened 2024-09-13
Chore/batched image upload and minor improvments

JohanHjelsethStorstad:
This branch allows for uploading multiple images at a by having the client batch images. The client also gets nice feedback of the batches upload status.

This PR also removes uniqueness of image names.


theodorklauritzen (approved):
Looks good! I think in the future we only need one upload form, and handle many and single file upload in the same form. In addition the drop zone for files should be larger. Maybe the whole popup.

JohanHjelsethStorstad:
Agreed. But the whole admin panel is a mess really

PR #321, opened 2024-09-13
Refactor/OmegaId

theodorklauritzen:
This reduces the size of the OmegaID QR code a lot. However the solution is maybe not that elegant.


src/services/omegaid/compression.ts
```diff
@@ -0,0 +1,109 @@
...
+        const payload = decompressPayload(rawDataSplit[0])
+
+        const signature = bigIntToBase64(rawDataSplit[1])
+
+        return {
+            success: true,
+            data: `${headerB64String}.${payload}.${signature}`,
+        }
+    } catch (e) {
+        console.error(e)
```
JohanHjelsethStorstad, line 105: console.log

src/services/omegaid/compression.ts
```diff
@@ -0,0 +1,109 @@
+import { JWT_ISSUER } from '@/auth/ConfigVars'
+import type { ActionReturn, ActionReturnError } from '@/actions/Types'
+import type { OmegaIdJWT } from '@/services/omegaid/Types'
+
+
```
JohanHjelsethStorstad, line 5: Maybe document the JWT -> QR compression on the top

src/jwt/parseJWTClient.ts
```diff
@@ -74,9 +74,6 @@ export async function parseJWT(token: string, publicKey: string, timeOffset: num
         const payload = readJWTPayload(token)
 
         if (!(
-            typeof payload.usrnm === 'string' &&
```
JohanHjelsethStorstad, line 77: I think this file is wrongly placed. First I think the function should be called parseJWTClient and this parse is specific to omegaId so why is it placed with the general jwt stuff. I think this should be split: one part is. in jwt libary it verifies and parses jwt on client. One part decodes and then uses this function - this should be placed with omegaId functionality.

JohanHjelsethStorstad (changes requested):

theodorklauritzen:
Hmmm, the main branch had a bug with a circular dependency in Vevenadapter, this broke feide login in the development environment. The fix for now is to not use the method handler, this is against the convention, and we should maybe discuss a more permanent solution.

PR #331, opened 2024-10-01
Fix/reset-password not building

JohanHjelsethStorstad:
This PR cleans up the reset-password page to follow conventions but manly fixes a major issue. There is something I do not understand. It seams like VevenAdapter must be imported in some way in all pages. importing getUser is a way to solve this.

### DIFF
--- a/src/actions/auth/resetPassword.ts
+++ b/src/actions/auth/resetPassword.ts
@@ -1,15 +1,22 @@
 'use server'
-
-import { createZodActionError } from '@/actions/error'
+import { getUser } from '@/auth/getUser'
 import { safeServerCall } from '@/actions/safeServerCall'
-import { sendResetPasswordMail } from '@/services/notifications/email/systemMail/resetPassword'
-import { emailValidation } from '@/services/notifications/validation'
 import type { ActionReturn } from '@/actions/Types'
-
+import { emailValidation } from '@/services/notifications/validation'
+import { sendResetPasswordMail } from '@/services/notifications/email/systemMail/resetPassword'
+import { verifyResetPasswordToken } from '@/services/auth/resetPassword'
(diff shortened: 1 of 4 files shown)


src/actions/auth/resetPassword.ts
```diff
@@ -1,15 +1,22 @@
...
     const parse = emailValidation.typeValidate(rawdata)
     if (!parse.success) return createZodActionError(parse)
 
     return await safeServerCall(() => sendResetPasswordMail(parse.data.email))
 }
+
+export async function verifyResetPasswordTokenAction(token: string): Promise<ActionReturn<{
+    userId: number
+}>> {
+    getUser({})
```
Paulijuz, line 20: There should be comment for why this is here. Someone will see this in the future and get confused. E.g. `// For some reason the project doesn't build if getUser (or actually VevenAdapter specifically) isn't included for all actions. This is most likely a webpack issue and should probably be fixed in the future, but for now simply including it (even if it doesn't actually do anything) will work.`

Paulijuz (changes requested):
Looks good! My only suggestion is to add a comment explaining the issue. I have written a comment which you can copy paste inn if you don't want to write it yourself.

src/actions/auth/resetPassword.ts
JohanHjelsethStorstad: Well maybe, but this will be refactored soon anyways. Hopefully we will have no problems with the general Action HOF

Paulijuz: Of course the best solution is to fix the issue, but it doesn't hurt to have a comment as long as this is how it's done. It only takes a few seconds to add.

Paulijuz (approved):
LGTM!

PR #335, opened 2024-10-07
Chore/npm audit october 2024

JohanHjelsethStorstad:
This PR
- Upgrades to node 22
- Fixes the most critical package vulnerabilities.

We have had problems changing react-pdf version because of a missing pdfs-worker. I have now found one that works. It does depend on using Promise.withResolvers - a new method. This has two consequences:
- We should upgrade to node 22 (so the method is available server-side)
- Some older browsers might have a problem - I have created a fallback.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
     - name: set up node
       uses: actions/setup-node@v3
       with:
-        node-version: '20.x'
+        node-version: '22.x'
 
     - name: Install dependencies 
       run: npm ci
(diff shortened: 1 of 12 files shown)


JohanHjelsethStorstad:
node 22 will become LTS in late October so I see no problem in upgrading

JohanHjelsethStorstad:
For some reason I had to explicitly type two things in VevenAdapter too

Paulijuz (approved):
Works on my machine at least. Can i use says that withResolvers has about 84% global coverage so I'd say it's good to go.

Did you format the prisma files by mistake by the way? lol

JohanHjelsethStorstad:
I think it might have happened when reinstalling on node 22

PR #328, opened 2024-09-26
Chore/events migrations

JohanHjelsethStorstad:
This PR should be merged #324. It migrates the part of the events system already implemented in #324

PR #339, opened 2024-10-10
Chore/users migrations

JohanHjelsethStorstad:
DOBBEL OMEGA
This PR adds user migrations.

### DIFF
--- a/src/prisma/prismaservice/src/dobbelOmega/migrateUsers.ts
+++ b/src/prisma/prismaservice/src/dobbelOmega/migrateUsers.ts
@@ -18,6 +18,13 @@ export default async function migrateUsers(
 ) {
     const users = await vevenPrisma.users.findMany({
         take: limits.users ? limits.users : undefined,
+        include: {
+            StudyProgrammes: {
+                select: {
+                    years: true
+                }
+            }
+        }
     })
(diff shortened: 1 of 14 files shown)


Paulijuz:
Took a quick look now and it looks good. I'll do a proper review once #328 is merged and the conflicts are fixed.

JohanHjelsethStorstad:
@Paulijuz. - now rebased on main

src/prisma/prismaservice/src/dobbelOmega/migrateUsers.ts
```diff
@@ -0,0 +1,379 @@
...
+                bio: user.bio ?? '',
+                acceptedTerms: undefined,
+                sex: sexMap[user.sex],
+                allergies: undefined,
+                mobile: undefined,
+                emailVerified: undefined,
+                createdAt: user.createdAt,
+                updatedAt: user.updatedAt,
+                imageId: vevenIdToPnId(imageIdMap, user.ImageId),
+                archived: user.archived,
```
theodorklauritzen, line 92: I don't think it is necessary to archive PN users.

theodorklauritzen (approved):
Great work!

It's probably nice to store if users are archived. However I don't think it's necessary to archive users. I think they still should be able to login, but with limited permissions

PR #345, opened 2024-10-17
Refactor/lib and containers folder

Paulijuz:
It's getting quite busy in the src folder and the root of the project so i created two new folders to declutter.
-`src/lib`: This is a folder I have seen in several next js projects. Here we can put all the library (or library like) code here. As a rule of thumb, if the code you're writing can be pulled out of the project and be reused in another project without any changes it can be put here.
-`containers`: I think the name is self explanatory. Just put all containers which we need inside here. I think prismaservice should be here too, but since it will be removed/reworked in the future i wont bother with it now.


theodorklauritzen (approved):
LGTM

PR #346, opened 2024-10-18
Feat/User Profile Admin Pages and Dots 1

JohanHjelsethStorstad:
This PR changes the profile UI by adding a profile view for the logged in user with links always available on all pages. It also refractors the user admin pages.

Then this PR implements the first part of the dot system. It is implemented using an expire date cached on the dot. This is a departure from omegaweb where this attribute was calculated using creation date and context from other dots. We. could discuss this more, but as I see it it makes reading way easier. However, the deletion of dots involves some logic now, and will be implemented later together with freeze periods. A new general system for freezing dots expiry in chosen time periods.


theodorklauritzen (approved):
I think that configVars.ts only should contain configuration variables, and not includers.This should be fixed at a later time. In addition I hope the expiry system will work well.

PR #347, opened 2024-10-18
Feat/Interest Group Pages

JohanHjelsethStorstad:
This PR adds a page to display interest groups and makes the interest-group service actually usable. It also
- Adds an alternative dynamicFieldsAsync fetcher of dyn. fields for an author to ServiceMethod. 
- Makes sure sessions (jwt) are invalidated for users when memberships change.
- Adds a new Auther for having a. permission or being active member of a group. This can be reused later for authoring committees.


theodorklauritzen (approved):
LGTM

PR #364, opened 2024-11-22
Feat/Shop

theodorklauritzen:
- Add Shop backend
- Admin page to administrate shops and products
- API to interact with shops
- A table that acts as receipts for product purchases.
- Connect student cards to users

Also see [API docs](https://github.com/vevcom/projectNext/wiki/API).

Remaining features:
- Connect to the money system
- Auth API KEY to only specific shops
- Visibility auth to how can edit shops, so users only can only access their shops
- Web shop
- OmegaID shop / Vandrebar


src/services/shop/Authers.ts
```diff
@@ -0,0 +1,16 @@
+import { RequirePermission } from '@/auth/auther/RequirePermission'
+
+export const ReadShops = RequirePermission.staticFields({ permission: 'SHOP_READ' })
```
Paulijuz, line 3: Not directly related to your PR specifically, but should we create an auther for each service method, even if the requirements are exactly the same? This seems like a lot of work (repetition) for little benefit?

src/services/shop/product/update.ts
```diff
@@ -0,0 +1,50 @@
...
+            ShopProduct: {
+                update: {
+                    where: {
+                        shopId_productId: {
+                            shopId,
+                            productId
+                        },
+                    },
+                    data: {
+                        price: data.price,
```
Paulijuz, line 43: If I remember correctly, the old vev stores the history of prices for a product. Not that it is necessary for our implementation, but it's something worth mentioning.

src/prisma/schema/user.prisma
```diff
@@ -62,3 +64,9 @@ model FeideAccount {
   userId      Int      @unique
   user        User     @relation(fields: [userId], references: [id], onDelete: Cascade, onUpdate: Cascade)
 }
+
+model registerStudentCard {
```
Paulijuz, line 68: Two small things.
1. I had to do a double take on what this model was intended for. Maybe a better name would help? Something like "StudentCardRegistrationQueue" or something along those lines.

2. UpperCamelCase is the convention for model names.

Paulijuz (changes requested):
Looks awesome. The implementation of the shop looks nice. I added a few minor comments. There are a few major things however that will be necessary to change.

Firstly, I overhauled the ServiceMethod implementation to simplify some things. I would like to merge that PR (#381) before this. All calls to the "ServiceMethod" function will need to be changed. This should not take a lot of time, but I can help you with this if you want, just let me know.

Second, with the overhauled ServiceMethod I have a simple testing framework set in place. It'd be nice to write some tests for this feature. But, since it's not merged into main yet we can also do this in a later PR if you want.

Paulijuz:
Also, great to see the wiki being updated!

src/services/shop/Authers.ts
JohanHjelsethStorstad: I prefer it for splitting purposes. Since authers are both used in frontend and backend if one method changes its author it will be reflected on both frontend and backend checks

JohanHjelsethStorstad: One auther for every service method that way a change will have no unintended side effects.

.eslintrc.json
```diff
@@ -432,7 +432,7 @@
         // limits the number of parameters that can be used in the function declaration.
         "max-params": [
             "error",
-            4
+            5
```
JohanHjelsethStorstad, line 435: What was this changed for?

src/services/prismaCall.ts
```diff
@@ -16,6 +16,10 @@ export async function prismaCall<T>(call: () => Promise<T>): Promise<T> {
     try {
         return await call()
     } catch (error) {
+        if (error instanceof ServerError || error instanceof Smorekopp) {
+            throw error
+        }
+
```
JohanHjelsethStorstad, line 22: Thanks for finding this very bad bug after changing to ServiceMethods

src/app/admin/shop/[shop]/EditProductForShopForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export function EditProductForShopForm({
+    shopId,
+    product,
+}: {
+    shopId: number,
+    product?: ExtendedProduct,
+}) {
+    const submitAction = product
+        ? updateProductForShopAction.bind(null, { shopId, productId: product.id })
```
Comment (Paulijuz) [inline src/app/admin/shop/[shop]/EditProductForShopForm.tsx:18]: with the new PR i created the utility functions `bindParams` and `bindData` for better readability

src/services/shop/authers.ts
```diff
@@ -0,0 +1,21 @@
...
+export const readShopsAuther = RequirePermission.staticFields({ permission: 'SHOP_READ' })
+export const readShopAuther = RequirePermission.staticFields({ permission: 'SHOP_READ' })
+export const createShopAuther = RequirePermission.staticFields({ permission: 'SHOP_ADMIN' })
+export const updateShopAuther = RequirePermission.staticFields({ permission: 'SHOP_ADMIN' })
+
+export const readProductAuther = RequirePermission.staticFields({ permission: 'PRODUCT_READ' })
+export const createProductAuther = RequirePermission.staticFields({ permission: 'PRODUCT_ADMIN' })
+export const updateProductAuther = RequirePermission.staticFields({ permission: 'PRODUCT_ADMIN' })
+
+// export const CreateProductForShop = RequirePermission.staticFields({ permission: ['PRODUCT_READ'] })
```
Paulijuz, line 13: should this be uncommented?

Paulijuz (approved):
Looks great!

Only two really minor things, but I have approved anyways as they are _really_ minor.

PR #368, opened 2024-11-25
Feat/Connect users

theodorklauritzen:
This connects existing users to users logging in. In the Veven adapter, I have removed the calls to service methods since this caused a wired initialization bug. I can explain this before we merge.
- Small bug fixes with the registration
- Added some extra details around the OmegaID QR code


src/app/(auth)/register/page.tsx
```diff
@@ -19,41 +19,37 @@ export default async function Register({ searchParams }: PropTypes) {
...
         if (!verify.success) {
-            console.log(verify)
-            return <p>Token er ugyldig</p>
+            redirect('./register')
         }
 
         if (user && verify.data.id !== user.id) {
-            // TODO: Logout
-            console.log('Should logout')
+            // throw new Error('Email is verified. Cannot continue registrations while another user is logged in.')
```
Paulijuz, line 32: Should this error be commented out? If yes then a comment explaining why,and wheter/when it should be uncommented would be nice.

src/services/notifications/email/systemMail/verifyEmail.tsx
```diff
@@ -15,7 +15,7 @@ export async function sendVerifyEmail(user: UserFiltered, email: string) {
         sub: user.id,
     }, emailValidationExpiration)
 
-    const link = `${process.env.DOMAIN}/register?token=${jwt}`
+    const link = `${process.env.NEXTAUTH_URL}/register?token=${jwt}`
```
Paulijuz, line 18: Really minor thing, but I'd consider depending on next auth specific env variables bad practice. If next auth changes the name of this env variable in the future it will cause a (minor) headache for someone in the future

Paulijuz (approved):
Looks good!

Just two minor comments, nothing I think you _need_ to change.

Paulijuz (approved):
Noice

PR #386, opened 2025-01-21
Chore/Licenses

JohanHjelsethStorstad:
This PR adds a dynamic license system for images


src/services/images/create.ts
```diff
@@ -1,61 +1,118 @@
...
+ * Creats  bad image this should really only happen in worst case scenario
+ * Ads it to the standard collection
+ * @param name - the name of the image
+ * @param config - the config for the image (special)
+ */
+export const createBad = ServiceMethodHandler({
+    withData: false,
+    handler: async (prisma, { name, special }: { name: string, special: SpecialImage }) => {
+        const standardCollection = await readSpecialImageCollection('STANDARDIMAGES')
+        logger.warn('creating a bad image, this should only happen in extreme cases.')
```
Paulijuz, line 96: i think that you could use a more descriptive word than "bad" here, from reading the doc i have no idea what a "bad" image is

also you misspelled create

src/prisma/prismaservice/standard_store/images/machine.jpeg
Paulijuz, line 1: Isn't this a bit larger than necessary? 12 mb seems like quite a lot for its use case

same goes for some of the other images, like "hovedbygget.jpeg", and "realfagsbygget.jpeg"

src/lib/feide/ConfigVars.ts
```diff
@@ -1,3 +1,3 @@
 import 'server-only'
 
-export const feideScope = 'openid email groups-edu userid userinfo-mobile userinfo-name longterm'
+export const feideScope = 'openid email groups-edu userid userinfo-mobile userinfo-name'
```
Paulijuz, line 3: What is the reason for this?

default.env
```diff
@@ -1,5 +1,6 @@
 # Server info
 DOMAIN=omega.ntnu.no
+SERVER_LINK_PREFIX=http://localhost:80
```
Paulijuz, line 3: server link prefix? maybe base url/domain, host url/domain would be better?

Paulijuz (changes requested):
Looks awesome! Proper licence information will be very nice to have. From some very quick testing everything appears to work without issues.

Just commented on a few minor things which I think can be addressed rather quickly.

Not really related to the PR, but we need to agree with whether or not we should bind before call on actions (i.e. `someAction.bind(..., data)()`). With the direction things are going this seems a bit redundant and unmaintainable so I would advise against it.

src/prisma/prismaservice/standard_store/images/machine.jpeg
JohanHjelsethStorstad: They are resized on seed though. The original size is never used

src/lib/feide/ConfigVars.ts
JohanHjelsethStorstad: This was not me I think this comes from shop PR?

Paulijuz (approved):
LGTM!

PR #387, opened 2025-01-24
Chore/next 15

JohanHjelsethStorstad:
This PR does a manual upgrade to next 15. Since Next has changed the params and searchParams api some changes were necessary. It also migrates certain pages to use the QueryParams api for safety and consistency.


JohanHjelsethStorstad:
@theodorklauritzen some of the registration pages have been modified - can you look through it?

Paulijuz (changes requested):
Everything seems to _work_ from what testing i did. However, I have experienced a few client side errors:

- On the login page I get a client side/server side hydration mismatch error related to some of the input elements. Specifically that some ids don't match.
- "Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>." on pages which use option with default value. For example create job ad form.
- "Each child in a list should have a unique "key" prop. Check the render method of `SlideSidebar`." When viewing the admin settings page.
- "In HTML, <th> cannot be a child of <thead>." Specifically for the admin "studieprogram" table, but this might apply to more tables.

I think all of these issues have existed before this PR was created, but Next js 15 has started throwing errors for these issues now.

JohanHjelsethStorstad:
Added some fixes now to fix the errors noted by @Paulijuz, but there are probably more, however, these have always been problems and I do not think they will break our production code

Paulijuz (approved):
LGTM!

Managed to find some other errors, but as you say they're not critical

PR #388, opened 2025-01-26
Refactor/docker base config

Paulijuz:
When writing tests it became clear that having three different docker files is hard to maintain as all of the shared services have to specified three places. Especially when the env variables to the pn container would be annoying to maintain as changing it would mean changing it three places. This PR fixes it by adding a third (fourth with tests) docker compose file (`docker-compose.base.yml`) which contains all the shared services. Then the specific compose file (dev, test or prod) uses the `extends` keyword to use the service from the shared file and applies the required modifications, for example exposing different ports or using different volumes.

This PR also adds a skeleton for the docker file for tests. The testing framework itself will be implemented in another PR.

Also, I have added the npm run commants `docker:dev`, `docker:test`, and `docker:prod` for simplifying starting up docker. I don't expect people to use them, but they might be useful for some newcomers if they find it hard to remember the exact command to start the project. The commands for dev and test also automatically copy the `.env.default` file.


theodorklauritzen (approved):
LGTM

JohanHjelsethStorstad (approved):
LGTM

JohanHjelsethStorstad (approved):

PR #391, opened 2025-01-27
Add scripts to make a user admin from the command line.

theodorklauritzen:
This is very useful for the first user on the system. Maybe we need another solution longterm.

In addition I fixed a bug in `apiHandler` at endpoints that don't expect any data.

### DIFF
--- a/src/app/api/apiHandler.ts
+++ b/src/app/api/apiHandler.ts
@@ -46,15 +46,22 @@ export function apiHandler<
     // TODO: I think I will rewrite this to be easier to read
     return async (req: Request, { params: rawParams }: { params: RawParams }) =>
         await apiHandlerGeneric<Return>(req, async session => {
-            const rawdata = await req.json().catch(console.log)
+            let data
 
-            if (!serviceMethod.dataValidation) {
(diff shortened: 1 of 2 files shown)


src/app/api/apiHandler.ts
```diff
@@ -46,15 +46,22 @@ export function apiHandler<
     // TODO: I think I will rewrite this to be easier to read
     return async (req: Request, { params: rawParams }: { params: RawParams }) =>
         await apiHandlerGeneric<Return>(req, async session => {
-            const rawdata = await req.json().catch(console.log)
+            let data
```
JohanHjelsethStorstad, line 49: I guess this will be removed soon

JohanHjelsethStorstad (commented):
I am not sure I am comfortable having a lot of random python scripts. It makes the project hard to maintain

src/app/api/apiHandler.ts
JohanHjelsethStorstad: Nope my point is there is no need to type validate soon

theodorklauritzen:
We could seed a admin user to prod that would also work fine. It's just nice to have a easy way to get admin permissions in prod

PR #394, opened 2025-02-09
Feat/tests

Paulijuz:
Tests are a bit rough around the edges right now as all the tests are written with raw prisma call. However, it is more important to begin writing tests now that major features are being completed than it is to wait for a perfect test system. Also, I think we can figure out our style for writing tests organically as we write them.


Paulijuz:
Created a draft PR to test tests gh action

theodorklauritzen (approved):
This looks great and incredibly needed.😍

PR #398, opened 2025-02-20
chore/Add Dockerfile to dependabot

theodorklauritzen:


Paulijuz (approved):
LGTM

Just a few questions:
- Why are we ignoring node 23?
- Should the node version arg/a node version env variable be used other places too?

PR #407, opened 2025-03-06
Refactor/lockers use service method

Paulijuz:
Things that are still missing:
- Translate validations to schema
- Figure out how to do authing with both bypass permission, normal permission, user id and group id


AndreasVJ (approved):
Looks good

PR #406, opened 2025-03-03
Refactor/services validation

theodorklauritzen:
This PR refactors validation and all service methods to use the ServiceMethods function wrapper.

PR #411, opened 2025-04-12
Feat/Cabin booking

theodorklauritzen:
MERGE #406 first!

This PR adds basic functionality for Cabin booking. I think merging this basic functionality before the branch grows too much is smart. In addition, I think I needs some confirmation from HeutteCom on the next steps.

## This includes:
- Database schema (Maybe the cabinGuest model can be generalized? Maybe the email should be unique)
- Price periods
- Release periods
- Booking page (Both for users and guests)
- Simple table to show bookings for admins
- Admin page to add products (Needs better UI to separate price periods, edit products, and remove products)

## Missing features:
- Admin functions to administer bookings (change existing, add new bookings for users, add booking outside the release period)
- Way to show your own bookings
- Sending confirmation emails
- Connecting event booking to events
- Change the cabin calendar to show how many beds are available if there is a bedbooking
- More function later in the process of a cabintrip

## Concepts
Price period is a period with a set of prices. This allows for updating prices often due to inflation.

ReleasePeriod is a period when it is possible to create a booking.


Paulijuz (dismissed):
Looks great! From som very simple testing the pure functionality seems to be working.

Just a few comments:
- For the cabin booking service methods i see a lot of duplication with the "no user" and "with user" version. Wouldn't it be possible to have the user be optional? If the zod schema is the problem i would recommend looking up "discriminated unions" in the zod docs. You could for example have an option "guest user" which allows passing in contact info, and another option "veven user" (or maybe another name) which just takes in the user id.
- I saw your race condition todo. This can be easily handled by wrapping the entire thing in a transaction and doing the entire thing optimistically. (https://en.wikipedia.org/wiki/Optimistic_concurrency_control)
- I see a lot of stray console log lol

Since all of these can be fixed at a later date i think its best to merge now as is. The PR is large enough as it is.

Paulijuz (approved):
LGTM

Same comments as before

PR #412, opened 2025-04-13
Refactor/seeder

theodorklauritzen:
MERGE #406 first!

This PR restructures the seeder, by removing prismaservce. This enables more shared code since the seeder is in the same npm project. This also makes use of how Prisma usually seeds data with the `prisma db seed` command. In addition the migration script is now a command, which I think makes more sense than a container.

My hope with this was to integrate the seeder with the tests. However, this seems a bit hard because Jest does not understand ES6 that well. But I think the PR is still a step in the right direction, I hope this doesn't ruin some sort of plan to extend prismaservice.


package.json
```diff
@@ -11,10 +11,17 @@
     "test": "jest",
     "docker:dev": "docker compose -f docker-compose.dev.yml up --build",
     "docker:test": "docker compose -f docker-compose.test.yml up --build --abort-on-container-exit --exit-code-from projectnext --attach projectnext --no-log-prefix",
-    "docker:prod": "docker compose -f docker-compose.prod.yml up --build"
+    "docker:prod": "docker compose -f docker-compose.prod.yml up --build",
+    "dobbelOmega-pull": "npx prisma db pull --schema ./src/prisma/vevenSchema/schema.veven.prisma",
```
Paulijuz, line 15: a really tiny thing, but i think i'd prefer a colon in stead of a hyphon here. the convention i tried to follow is `[category]:[command]`. the "dev-seed" is correct tho as it does both

Paulijuz (approved):
this one sparks joy :D

PR #413, opened 2025-04-13
Refactor/user selection level

theodorklauritzen:
MERGE #406 first!

This PR adds different levels of fields to expose from the user object. This should hopefully reduce the amount of user data that is exposed.

### DIFF
--- a/src/actions/mail/read.ts
+++ b/src/actions/mail/read.ts
@@ -7,7 +7,7 @@ import { readMailAliases } from '@/services/mail/alias/read'
 import { readMailingLists } from '@/services/mail/list/read'
 import { readMailAddressExternal } from '@/services/mail/mailAddressExternal/read'
 import { getUser } from '@/auth/getUser'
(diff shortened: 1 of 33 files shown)


JohanHjelsethStorstad (changes requested):
I think it is more correct to use Prisma giving it the selection to generate the types, however, I still think the types for all different selections is sort of cluncy. Ideally I think all reads should have a filterSelection (an includes) to be as explicit as possible. -> this filter has a name (I belive we probably should call them selections as it is consistent with SELECT) and this name always corresponds to a type by Prisma.UserGetPayload. This type should have the same name... . I am sort of wondering, why do we need the type? can't we have a wrapper takin in the selection object you want to use. So if you need the UserNameFiltered type you use SmartWrapperName<typeof UserConfig.filterNameSelection> or with som name changing Selection<typeof UserConfig.selections.name>

theodorklauritzen:
Closing, since this has been inactive for a long time

PR #414, opened 2025-04-24
fix: service method .execute has proper type

JohanHjelsethStorstad:
This PR fixes an issue with ServiceMethod .execute expecting wrong type when a schema transforms a value.


theodorklauritzen (approved):
LGTM :))

PR #417, opened 2025-04-30
Feat/landing page

theodorklauritzen:
I need some feedback, I have some breaking changes that needs to be disgussed.

Additions:
- Added seeding of jobads, and fixed the seeding of news
- Change all `import 'server-only'` to `import '@pn-server-only'`. This will make other scripts be able to use our existing server code. This means that we no longer want to write `import 'server-only'` any place. This way the servicemethods can be used by the seeder, and hopefully jest
- Update the layout of the image card
- Added two separate pages for `/`. The one when you are logged out is not changed, and when you are logged in, you'll see news, events, and jobsads.


Paulijuz (approved):
Nye forsiden er :fire:. Mangler litt polish, men det kan fikses senere 

En ting som kan være verdt å tenke på er at seederen ikke har brukt service methods før nå. Ideelt sett så burde hele greia fungere på samme måte, så kanskje det må sees på i fremtiden? Eventuelt kan vi bestemem oss for å ikke bruke services methods i seederen for å holde det likt med det som allerede er.

Fin løsning med `server-only` problemet.

PR #419, opened 2025-05-06
Test/integrate seeder

theodorklauritzen:
MERGE #417 FIRST

Integrate the seeder into the tests


Paulijuz (changes requested):
Very nice!

Only thing is that the tests fail on my machine because the seeder takes to long. You can increase the time out with something like this:
```ts
// setup.ts
beforeAll(
    async () => seed(false, false, false), 
    30 * 1000, // Set timeout to 30 seconds
)
```

We should probably reconsider what we seed in the future as it takes a solid 20 seconds to run the most basic test on my machine now, but it is fine for now.

Paulijuz (approved):
LGTM

PR #422, opened 2025-05-15
Feat/event registration

theodorklauritzen:
Merge #419 first!

Add functions to register for events.

This is missing authing with the visibility system, and a way to pay using the money system.

I also changed how Dobbel Omega for users are working. This new aproach uses a object to keep track of the users. This makes it easier to migrate with some limits, and make sure all dependencies are resolved.


Paulijuz (approved):
From some short testing everything seems to work as expected.

Some minor hydration errors with `In HTML, <span> cannot be a child of <tbody>.` on the events page. But these can be fixed at a later date togheter with a round of polish.

There also seems to be some issues regarding the registration and waiting lists not updating after changes. Again, fix at a later date.

LGTM!

PR #425, opened 2025-05-29
Feat/applications phaestum cd

JohanHjelsethStorstad:
This PR includes the beginnings of the application system
- Models for applications
- CRUD applications and periods
- CD to application deadline

Missing:
- UI polish
- admin so APPLICATION_ADMINs can see and moderate all applications and committeemembers can see their applications.

I suggest adding a q-param to /applications/[periodName]/?username=... so an admin can view applications of any user not only its own username, and more importantly the route /applications/[periodName]/[commiteename] where admins and committee members can see all applications to their committee.


JohanHjelsethStorstad:
Also the /applicationperiods has been added to the nav-menu, however if an application period is active it should be linked directly in the navbar - this must be checked on any render of the main layout

src/services/applications/methods.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    export const destroy = ServiceMethod({
+        paramsSchema: z.object({
+            userId: z.number(),
+            commiteeParticipationId: z.number()
+        }),
+        auther: ({ params }) => ApplicationAuthers.destroy.dynamicFields({ userId: params.userId }),
+        opensTransaction: true,
+        method: async ({ prisma, params }) => {
+            prisma.$transaction(async (tx) => {
```
theodorklauritzen, line 205: I don't think we should allow users to withdraw their application after the application deadline, so there should maybe be a check for that here.

src/services/applications/periods/methods.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    export const read = ServiceMethod({
+        auther: () => ApplicationPeriodAuthers.read.dynamicFields({}),
+        paramsSchema: z.object({
+            name: z.string()
+        }),
+        method: async ({ prisma, params }) => prisma.applicationPeriod.findUniqueOrThrow({
+            where: { name: params.name },
+            include: {
+                committeesParticipating: {
```
theodorklauritzen, line 23: In the config file a includer i defined that matches this includer.

theodorklauritzen (approved):
Great work, some minor details I commented on, but it is not that important.

theodorklauritzen (approved):
LGTM

PR #445, opened 2025-09-24
Chore/update scss

theodorklauritzen:
This PR fixes a lot of the different warning that fills up the logs


Paulijuz:
I see that some warnings still pop up. Is this a WIP or should we just fix those later?

PR #464, opened 2025-10-23
Refactor/cms subservice

JohanHjelsethStorstad:
This is a big PR that uses the new subOperation system to refactor the cms and implement it as part of the different systems that need it.


Paulijuz (approved):
Amazing! This one has been a long coming and it's really a big milestone to have this finally done. I feel you came up with one of the cleanest possible solutions given our architecture.

I don't have anyspecific requests for this PR, its large enough as it is, but I have some general thoughts for the future.

I feel like `authorizer` and `ownershipCheck` could be one function. I feel like the division is a bit artifical and from what i can see they are just used in conjunction within the subservice implementation anyway. I think I see why it was done this way, but I feel like it the need for this split stems from a weakness of the current implementation of the Auther system rather than any actual logical requirement.

I feel like SubService operation has quite many (too many?) different generic and parameters. The whole implementation feels way more complex than what it actually does. I don't especially like `implementationParams`, it kinda breaks the whole point of an abstraction in that the highest level shouldn't need to know about how the lower levels work. I see why it would be difficult to implement and if i understand you correctly you gave it a shot to combine both implementaitonParams and params? However i still think we should give it another shot as it would make the interface simpler. I can take a look at it if you want? I don't think it would be to hard to change later as it would just be a merging of implementation params and params, the logic would stay the same.

Not really related to what you have done, but I think it might be beneficial to export `baseSchema` from `schemas.ts`. I have experienced situations where its nice to reference some fields of another model and then it makes more logical sense to reference the base rather than one of the derived schemas. This would of course entail that we should have more descriptive names than "base", but code quality wise i think that would also be an improvement. 

Another thing is that i never really liked the word "special". With this PR we have an opportunity to change that. I really like the "special" cms system, but the word is not really descriptive of what it does. For example, read one of your comments

>  Check if a link with id is special with special at[t]ribute in the provided special array.

Wouldn't "specialPurpose", "uniquePurpose", "systemPurpose", just "purpose", "reservedFor" or "systemUseCase" be for example better? Try reading again:

> Check if a link has one of the special purposes in the provided array of special purposes.

If you want to keep "special" that's fine, but then we should clearly explain somewhere what the heck a "special" article is. But in my opinion, most people would find it easier to understand that an article has a special/unique/system purpose rather than it being special and having a special attribute.

Again, just food for thought and none of this needs to be handeled in this PR. I thinks its better to just merge it as is.

Finally, I just say that its always nice to see comments here and there. Especially for such complicated systems. <3

LGTM!

Paulijuz:
Ah yes, I see now, my bad. From skimming the code i really thought that ownershipCheck was that the user owned the given resource. Then I agree that they should stay separate, but also that it definitively  needs another name.

Paulijuz:
> It is hard because, well, the special parts are used for many different things.

Would you say that those many different things are many different _purposes_?

Paulijuz:
We should probably stop discussing here as this PR is closed, we can continue the discussion on discord or wherever if you want

PR #472, opened 2025-11-03
Chore/clientside sessions

JohanHjelsethStorstad:
This PR removes getUser and useUser to move to new authers and authresults.

To remove getUser some services now do not have auth checks but the auth object has been set up ready for their refactor.


Paulijuz:
Looks really good. Just one small thing. I never really liked the word "auther". I always read "author", its non standard from what i can see, and it only saves four letters compared to the much clearer "authorizer".

I feel like for new people looking over the code, seeing something like this
```ts
const canEdit = useEditMode({ auther })
```
will look like a spelling mistake.

May i suggest switching to "authorizer" now that we have the chance? I have done this in the branch `chore/clientside-session-authorizer` so that you can have a look.

JohanHjelsethStorstad:
I agree. We can change it. Keeping names short has no purpose in my opinion. Just one question: the property authorizer often refers to some different things now: are the dynamic and/or static fields bound. And in the definition of a ServiceMethod or implementation of SubServiceMethod it is a function returning an authorizer with the fields bound. It is probably not a big deal since the type has the information, but there is some disconnect when the types have different names Authorizer AuthorizerStaticFielsBound and so on but the prop is always authorizer.

Paulijuz:
I totally agree. The naming could be a lot better, but this is one step in the right direction at least.

Paulijuz (approved):
LGTM

Im guessing this wont actually work

Issue #439, opened 2025-09-08
Feat/koileskap leaderboard


theodorklauritzen:
Add a button on the profile page to participate in the frigde leaderboard.

PR #473, opened 2025-11-03
Feat/Consent form fridge leaderboard

Sherine7734:
being shown in the Kioleskap Leaderboard.

fix: #439

### DIFF
--- a/src/prisma/schema/user.prisma
+++ b/src/prisma/schema/user.prisma
@@ -28,7 +28,6 @@ model User {
   memberships       Membership[]
   credentials       Credentials?
   feideAccount      FeideAccount?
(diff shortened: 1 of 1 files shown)


src/prisma/schema/user.prisma
```diff
@@ -28,6 +28,7 @@ model User {
   memberships       Membership[]
   credentials       Credentials?
   feideAccount      FeideAccount?
+  kioleskapLead     Boolean             @default(false)
```
JohanHjelsethStorstad: I think this field name is non-descriptive. It should at least include the word consent. Also: avoid abbreviating the name.

Paulijuz:
This has become stale and as such I will close it until it is ready again

PR #534, opened 2026-05-15
Feat/logger

### LINKED ISSUE #101: feat/logger
A form of logging is needed. I think we can log all database actions quite easily as prisma has a built in logger we can output to where we want. However, we should probably log other stuff too. And too do that we need a standardized way (a library or something) to use for logging.

HaakonMikalsen:
Add logging support
Add automatic logging to database events


Paulijuz (approved):
LGTM 🎉

PR #551, opened 2026-07-29
Feat/image svgs and async uploads [OWTR #3.1]

JohanHjelsethStorstad:
This is PR # 3.1 in the one week till realese series. These should be merged sequentially

This PR adds support for for svgs. Now that image collections are a subservice we have the ability to require the committeelogos collection to only support svgs.

One problem that became painful in #548 was that uploading was slow and gave long prisma transactions. Now we have a separate service running in the background resizing and converting images. While this is not finished we serve a blur. Also added a new image size now that resizing does not cause time problems.


src/lib/images/worker.ts
Paulijuz, line 1: The name "worker" is extremely generic. Should be changed to "image-worker" or "image-resize-worker" or something like that in the future.

Paulijuz (dismissed):

Paulijuz (approved):
hell fucking yeah

PR #552, opened 2026-07-29
Feat/news auth [OWTR #3.2]

JohanHjelsethStorstad:
This is PR # 3.2 in the one week till realese series. These should be merged sequentially

Simply implements use of visibility for news, (It was not authed before)


kake21 (approved):
Looks good, functions as intended. Remember to auth before release 👍 💯

